PR #91, opened 2015-06-09
Fix issue with servers that have GSSAPI authentication enabled.

lsowen:
Force SSH client to only try publickey and password authentication modes (in that order).


zeapo:
nice! That should, indeed, solve the problem.

jelmer:
Thanks!

travisjayday:
How can you fix this though? The server I'm connecting to requires GSSAPI authentication so I can't just disable/circumvent it... :-(

jelmer:
@travisjayday How do you obtain Kerberos credentials on Android?

travisjayday:
I'm trying to connect to an ssh server that uses GSS with kerb5. Since it turns out the Android jre doesn't have that implementation, I found that you can cross compile Kerberos source and write a wrapper for Android. Someone already did it. Still painful though :/

PR #99, opened 2015-07-01
Multi-select (for password deletion)

wongma7:
This replaces the popupmenu with an actionbar. It's based on: http://www.grokkingandroid.com/statelistdrawables-for-recyclerview-selection/ . I'm not sure where to put the ActionMode & ActionMode.Callback (I just started learning Android).


zeapo:
Thanks for the PR. I will try it this WE and get back to you.
From the code I've read, it looks like a working solution.

PR #101, opened 2015-07-11
Password Generator 2

wongma7:
When adding a password entry, press the new 'plus' button to open the dialog. Some things that still need to be done:
1. I suspect people who use the generator will want their preferred generation settings obscured. Currently I just use SharedPreferences with filename 'pwgen' and save preferences whenever a password is generated (only when 'Generate' is pressed). I'm not knowledgeable about this stuff...is using something like this https://github.com/scottyab/secure-preferences acceptable?
2. I ported pwgen directly, pretty much line-by-line, and some of its behavior doesn't translate perfectly to the simple dialog I made. i.e. if you check 'Symbols' it means 'MUST include at least one symbol' (pwgen -y) but the other options only mean 'MAY include x' (just pwgen or pwgen -n). Personally I'm okay with this but it might be a bit confusing & people might want more options (as in the other port, which I didn't bother with)?
3. The button looks a bit out of place.
Also, having my pwgen code as a library & dependency can be done if desired but I'm having trouble putting it up on jcenter.

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
+++ b/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
@@ -20,6 +20,7 @@ import android.view.View;
 import com.zeapo.pwdstore.crypto.PgpHandler;
 import com.zeapo.pwdstore.git.GitActivity;
 import com.zeapo.pwdstore.git.GitAsyncTask;
+import com.zeapo.pwdstore.pwgen.PRNGFixes;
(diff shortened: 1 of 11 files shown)


zeapo:
Amazing work!

One change and I will merge it. Could you put the default length to 20 and pronounceable to false?

Regarding secure preferences, that seems nice, but we should discuss it in a separate issue.

wongma7:
I accidentally pushed without splitting my commit but the change is there.

& in that case I won't bother with it unless somebody else asks.

Issue #43, opened 2014-11-21
Generate SSH key


Beanow:
For authentication I prefer to have an SSH key for each device.
It would be great if the application can generate a new key and share the public key.

zeapo:
If anybody wants to work on this, see : http://www.jcraft.com/jsch/examples/KeyGen.java.html

We already use jsch (via jgit).

tverlaan:
What about using this: https://play.google.com/store/apps/details?id=com.madgag.ssh.agent&hl=en as a solution for SSH keys.

zeapo:
Nice! That would definitely be useful for me too. We've got to check that they have good intent api for third party apps

zeapo:
@Beanow This feature is now merged.

PR #103, opened 2015-07-15
SSH key generator

wongma7:
For #43. New option in Settings 'Generate ssh-key' will create a private key .ssh_key in the same location an import would put one and will also create a public key .ssh_key.pub in that directory. The public key is shown to the user immediately after generation so that they can share it with their git server. Also, the dialog that appears when a user attempts git operations without having an ssh key (when ssh key is set as their authentication mode) has been changed to have 'import' & 'generate' options as opposed to just 'ok'/import.
Goal here is to handle & generate just one ssh key pair to make it easier for users to get things set up from the app. More complicated handling of keys I think won't be necessary until multiple repos are supported & hopefully another app to handle it will be available by then... 
To-dos:
1. Let the user export/see the public key whenever they want. Can this be an additional item in the settings? And the public key should be deleted/not shown if a new private key is imported & overwrites the original private key.
2. There are lots of skipped frames when 'Generate' is pressed.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -32,7 +32,7 @@ dependencies {
     compile 'com.jcraft:jsch:0.1.52'
     compile 'org.apache.commons:commons-io:1.3.2'
     compile 'com.jayway.android.robotium:robotium-solo:5.3.1'
-    compile 'com.melnykov:floatingactionbutton:1.2.0'
+    compile 'com.android.support:design:22.2.0'
     compile 'net.rdrei.android.dirchooser:library:2.1@aar'
     compile group: 'com.google.guava', name: 'guava', version: '18.0'
 }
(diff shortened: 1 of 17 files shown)


wongma7:
Okay, done I think. So there are now 'generate ssh key' and 'view generated ssh key' options. The latter option is always visible but enabled/disabled appropriately

zeapo:
Nice! That was a needed feature.

Do you think it is possible to support multiple pairs of keys? That way, once we implement the multiple stores mode, we let the user select which keys to use.

I play with it this WE, I merge then tag a new version with these awesome new features.

wongma7:
Sure, I can work on that. Now that I think about it, it should not be too difficult. I am thinking of a single 'choose ssh key' preference where you can choose/generate/import private keys or key pairs. Then simply store the name of the preferred key in a preference.

zeapo:
The passphrase should have a password style and a monospaced font. With a small checkbox to show the passphrase.

PR #117, opened 2015-08-12
Autofill

wongma7:
Addresses https://github.com/zeapo/Android-Password-Store/issues/50 on Android 5.0+ and is merely a convenience copy/paste feature for 4.3+. The aim is to offer something like what is in lastpass and dashlane. Right now it only fills app password fields and you can associate apps with only one password.

Mainly I still need to add instructions/shortcuts for enabling the accessibility service, the 'per-app settings' page is not done, and strings. Also, I'll try to make the dialog smaller, maybe by replacing text with icons.

Does the 'per-app settings' page make sense? The alternative to using search suggestions to add apps would be listing all apps and just filtering the list, like in lastpass.

It takes a bit of work to set up, so here is a gif of what it looks like right now http://gfycat.com/WarlikeAltruisticBarbet . I have a password called 'eBay'...so it matched that with the eBay app.


zeapo:
I will try it this weekend. The screencast looks good.

wongma7:
I've finished making my changes I think.

zeapo:
can you explain how it works exactly?

wongma7:
Ideally it would be intuitive enough that anyone could figure it out, but that is unlikely to happen. I suppose I should explain somewhere at least briefly 'once you enable the service, when you click/focus a password field, a dialog offering to fill the password field with the app's matched password (if any) will appear.' But the per-app settings are probably not clear either, I took the "matched" term for associating apps with passwords from lastpass

zeapo:
Sorry for the delay, I'm extremely busy lately.

Maybe it is time for us to add a tutorial screen :)

wongma7:
No problem. The latest dialog has some screenshots, which should help a bit... Do you mean like a fancy page/series of pages? that opens only the first time they try to enable this feature? Or maybe just one such tutorial for the whole app.

PR #121, opened 2015-08-21
Support for non-git stores

zeapo:
When creating a store, the user can chose to disable the following options:
- Git
- The use of internal memory, disabling it will let the user create the store on the sdcard
- The use of a remote store, disabling this will create a git repository with no origin


zeapo:
will be handled in V2

PR #123, opened 2015-08-28
get rid of the WARNING: Dependency org.apache.httpcomponents:httpclie…

betheg:
…nt...


betheg:
update travis.yml to use sdk 23.0.0 #124

PR #126, opened 2015-08-29
PgpHandler: fix possible npe.

betheg:
when you clicked openpgp_key_id_pref then findViewById(R.id.progress_bar_label) is null.


wongma7:
Thank you. (I'm being quite sloppy; totally overlooked key selection)

Issue #144, opened 2015-11-04
Autofill


JasperWeiss:
Hi, I enabled autofill but it doesn't seem to do anything. It should automatically match although I've no idea what is supposed to match with what since I'm not sure what is supposed to go behins the /. 
I also tried to manually select a password but when I go to any site and fill in my user name nothing happens.
How does autofill work?

wongma7:
Hi, sorry for the confusion. It doesn't yet work with websites. A major missing feature, but it's being worked on. Documentation also ought to be on the way. In apps, the automatic match will simply search for the app name in your stored passwords and if it doesn't find anything no popup will show.

JasperWeiss:
Great I got it to work with spotify,
It should include usernames. 
So that, I can store multiple accounts and pick one from the list. And then it should fill in the username automatically. That would be awesome!
Also, what keeps me from replacing all passwords with strong randomly generated ones is the fact that there isn't a nice client for other platforms like windows. So there's no way to login without my phone (ignoring the linux command line utillity this is modelled after)

wongma7:
Yeah, having usernames would be a great addition. The #1 priority for me atm is getting websites to work nicely; ability to quickly choose from multiple accounts will probably follow as a result.

https://github.com/IJHack/qtpass may be an option on windows, don't use it myself though.

JasperWeiss:
Great idea here.
What if, I could use my android device to fill in passwords on other devices.
Essentially, using root passwordstore could act as a keyboard over usb. So I could plug in my phone and have it paste the password like a ubikey does.
With usernames it could paste the username, tab, and then password.

This is especially great in case you're not on your own computer. Would it be possible?
Because that would take away the problem of not having my password on any other device than my phone. (Just typing it over is really quite inconvenient)

wongma7:
Interesting idea, i'm sure that's possible...but it's infeasible. For the phone to act like a USB keyboard it needs a modified kernel. I'm sure in any case it would be more feasible to just improve pass on windows.

Another possibility, if you're comfortable with your passwords going through the internet and somebody else's servers, is using something like pushbullet universal copy and paste... I assume that is unacceptable to the vast majority of this app's users, but worth pointing out if we're exhausting all possibilities...

JasperWeiss:
Having to install an application every time I use a computer at school will only make managing passwords harder.
Is there really no way to have a Ubikey like way to paste my logins onto computers without custom kernel?
[...]

Issue #139, opened 2015-10-21
Show autofill popup on all passowrd fields.


kevincox:
It would be nice if the autofill popup was shown on all password fields regardless of whether a match could be found. This is because if I am in a password field the answer is almost certainly in my password vault so it should give me an option to select which one I want to use. It would be awesome if after selecting what password to use I could select "Add this as a pairing" or "Ask me every time" so that depending on the app I could select each time or save that association.

wongma7:
Good suggestions. I will add an 'Ask me every time' option for always showing the autofill popup for all apps. Useful for setting up associations for the first time in an app. I am not sure if it should be the default setting though. (side note, the auto match is not too good to begin with and could be improved...)

As for being asked for which password to use...right now in the popup if you want to use a different password you have to tap Settings -> Match with... -> then find the other password -> OK -> then go back to the original app and fill the field. A better solution could be maybe a spinner of your saved associations in the popup, not sure if that's possible.

kevincox:
That sounds great. Setting up the associations is exactly what my use case is. Going back and forth between the two apps to do it isn't ideal.

Better auto-matching is always welcome but there are a bunch of cases where I wouldn't expect auto-matching to work no matter how intelligent. For example logging into the "Hearthstone" app requires the credentials for "battle.net" in my password store. Another example is apps that use "Facebook Connect" login in an in-app browser. The app can be anything but I want to use my "facebook.com" password.

Thanks for looking into this and if you need a beta tester I would be pleased to help you out.

Nis2:
This is the way to go as it would also allow filling password for web sites. AFAIK you can't right now because doing so would associate a password with the chrome app instead of the url.

kevincox:
As @nis2 said this would also fix https://github.com/zeapo/Android-Password-Store/issues/144

wongma7:
#146 is pretty much done, I am just doing some basic testing for a couple days then I suppose I should get it released for you guys to test

Issue #50, opened 2014-12-04
Android clipboard is vulnerable to hijacking


emilecantin:
As evidenced [here](http://stackoverflow.com/questions/22277598/permanently-listen-to-clipboard-changes), it is apparently possible for any app on the device to listen for changes on the system clipboard and to get its contents.

Another password manager, Keepass2Android, implements a keyboard to get around this problem. I think we should do the same.

c4software:
Hi,

Its a nice implementation, but if (like me) you use Password Store in a Chrome context, its seems not working (because its a Website not an app). The idea of an input method is still not an option ?

wongma7:
It is still a possibile option I think. I believe in the latest versions of Android switching keyboards is more convenient and usable too. I'll work on it but no eta

polyzen:
In kp2a(keepass2android), you can have it [automatically switch to the kp2a kb and back](https://keepass2android.codeplex.com/wikipage?title=Advanced%20usage%20of%20the%20Keepass2Android%20keyboard) (doesn't seem to work inside terminals, but eh).

Also, `pass` seems to use "username" as the standard field. `passff` uses "login", but that's configurable

JasperWeiss:
When using autofill, does it utillize the clipboard? It requested some sort of permisson for autofill so I assumed this was some sort of android 5.0 feature that circumvents the need for the (insecure) clipboard

zeapo:
Autofill does not use the clipboard and the recently added share button allows to share the password in clear text to another app if needed.

PR #146, opened 2015-11-07
Autofill: multiple associations per app & WebView/Chrome

wongma7:
https://i.imgur.com/uATEWyY.png
https://i.imgur.com/unCJwwP.png

Mentioned in #144, #139, #50 .
First you go to Chrome and add all the passwords for websites to it (press + button to add, and tap a chosen password in the 'matched' list to remove). Then they'll show up when you tap on a website password field. The dialog is small so you will have to scroll to find passwords and now you just tap the password instead of the old 'Fill' button.

Please test/comment!! I know it's not close to ideal but I want to get this functionality out there this weekend as I won't have time to work on it for a while.

Getting the url from Chrome and using it will be possible, I am pretty sure, so in the future I want per-website associations instead of stuffing everything under Chrome/the browser app (except in Android 6). But there is no way to get the url associated with a WebView.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ language: android
 android:
     components:
         - android-23
-        - build-tools-23.0.1
+        - build-tools-23.0.2
         - platform-tools
+        - tools
         - extra-google-m2repository
         - extra-android-m2repository
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,10 @@ language: android
 
 android:
     components:
-        - android-23
-        - build-tools-23.0.2
         - platform-tools
         - tools
+        - build-tools-23.0.2
+        - android-23
         - extra-google-m2repository
         - extra-android-m2repository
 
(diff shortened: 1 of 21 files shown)


zeapo:
Looks promising. Do you think we could test this using `uiautomatorviewer`?

wongma7:
Yes, I could write some basic tests to ensure a WebView is found when there is one in the window, the popup shows up when needed, and SET_TEXT actually works...that would technically "cover" it all. Actually this is a good opportunity because I have not yet tested WebView filling on devices under Android <5 and in fact it should not work.

BTW, it should be possible to get page title of a WebView so disregard negativity in my last sentence!!

wongma7:
This is about done I think, it uses the site URL and users can add settings for URLs. I think I confused uiautomatorviewer with uiautomator earlier. uiautomatorviewer shows the tree is kind of a mess, but getting to the WebView doesn't take much and the URL bar has a unique ID so I can reliably get it!

In older non-Chrome webviews it can't get the URL. But in such cases one can enable 'always show dialog' and then use 'pick password...' shortcut.

I will (try to) work on documentation for the whole app after this is finished.

(btw I don't know why travis is failing... :/ Anything special I need to do when upping the build tools?)

PR #189, opened 2016-06-10
Some restyling

DSIW:
* Password cards are restyled
* Password entry is restyled
* Counter will be displayed if some items are selected in list
* Password formular for new passwords is restyled
* Actionbar icons are updated

I know that these restylings aren't complete compatible with your [V2 Design](https://github.com/zeapo/Android-Password-Store/issues/131) but I think it could be a first version. More changes are done in the future.

What do you think about it?

See the screenshots below as impressions:

detail
list
list_selected
new_password


zeapo:
Nice! I'll try it and get back to you.

It could be indeed a first step into the v2 which is far from being ready.

zeapo:
I find it difficult to distinguish categories from passwords :/
Do you think putting bold on the passwords names would be an option? Or do you have another solution?
Maybe I got used to color differentiation :)

DSIW:
I tried bold typeface and some color changes, but that doesn't look nice. I don't think that is a big problem, because password files are at the bottom of the list and the icon can be used as type indicator.

zeapo:
It's OK, just personal preference :)

PR #201, opened 2016-07-23
oisafe2pstore.hs: script to migrate from OI Safe

crosser:
Quick and dirty Haskell script that converts OI Safe export CSV
format into the password-store repository format. After conversion,
you will need to `git add` the newly generated files, commit and
push them to the server. Then pull or sync on the Android device.

Signed-off-by: Eugene Crosser <crosser@average.org>


zeapo:
Thanks for the contribution.

PR #208, opened 2016-08-18
Fixed crash when ellipsing line.

danielvandenberg95:
See issue https://github.com/zeapo/Android-Password-Store/issues/207


zeapo:
Thanks for the PR.

Could you check if using `scrollHorizontally` with the actually present `maxLines : 1` does the job?
If I recall well, `singleLine` has some performance issues with scrolling. If `scrollHorizontally` does not fix the crash, then I will merge with `singleLine` with `maxLines : 1` removed.

danielvandenberg95:
Whoops, did not mean to add that commit...

zeapo:
you can squash your commits to have a single one

danielvandenberg95:
I'm not exactly a pro at github, but if I'm not mistaken it should be at the original state now. I've tried it with scrollHorizontally enabled,, and it still crashes. The weird thing is that Android Studio doesn't seem to mark it as deprecated in this use-case.

zeapo:
Thanks for the PR, I just squashed the commits.

You can do squashing in local with git see https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

PR #210, opened 2016-08-19
Added the ability to move passwords around.

danielvandenberg95:
Woohoo, managed to squash all commits into one :) Kinda...

But well, this pull request adds the ability to move passwords around within existing folders/categories. A future expansion would be adding folders/categories while moving the passwords around. A problem that would arise of that would be, how to handle empty categories?


zeapo:
Thanks for the PR.

This is a big one. I will take a look at it in depth ASAP.

danielvandenberg95:
https://github.com/zeapo/Android-Password-Store/issues/175

(Just mentioning as this would resolve part of this request.)

zeapo:
There is a lot of duplicate code for the adapters. Do you think it's possible to have a single abstract adapter where we just override its behaviour in both cases (moving and listing)?

danielvandenberg95:
Should be possible. In all honesty I think there is more duplicate code than just in the adapters, but I tried to stick with the existing structure as much as possible. To be honest, the existing structure is quite foreign to me. I'll see what I can do.

danielvandenberg95:
I've generified the adapters, and looked around for more duplicate code, but couldn't find too much that'd be worth generifying. If you do happen to run into any, just let me know and I'll look into it.

zeapo:
Nice, I will try it tonight and the next few days and get back to you before the weekend :)

PR #213, opened 2016-08-27
Share as PlainText

danielvandenberg95:
Added the ability to share a password as plaintext using any app, including ones like BluetoothToPc.


zeapo:
that looks good to me

PR #230, opened 2016-10-16
Added the ability to make an invisible password visible

danielvandenberg95:
Added the ability to make an invisible password visible using a button, and added some rough translations.

This allows you to hold down a button to see the password while it's being hidden. From personal experience, when using the password store in combination with https://github.com/danielvandenberg95/BluetoothToPc and with passwords set to invisible, this can be very convenient.


danielvandenberg95:
I've got the idea someone broke travis…

danielvandenberg95:
Can I force a travis re-run somehow?

zeapo:
AFAIK only if you push force or create a new commit, sry :/

danielvandenberg95:
http://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

Close and reopen apparently helps :)

PR #278, opened 2017-02-27
Added app shortcuts

ZidHuss:
Creates dynamic app shortcuts when decrypting passwords. Allows faster access to recently used passwords.

Signed-off-by: Hussein Mohammed <zidhussein@gmail.com>


zeapo:
Thanks for the PR.

Could you update `.travis.yml` by adding the sdk v25? See these lines: https://github.com/zeapo/Android-Password-Store/blob/master/.travis.yml#L7-L10

PR #290, opened 2017-04-15
Make passwords sort case insensitive

Tunous:
This makes the sort order behave like when using the `pass` command.

Before:

- BB
- aa

After:

- aa
- BB


zeapo:
thanks for the PR

PR #315, opened 2017-07-15
Ignore .extensions directory

felixb:
Ignore .extensions directory when traversing password tree.


zeapo:
Thanks for the PR; 
Seems like we have duplicate code, so could you add the same bit it at https://github.com/zeapo/Android-Password-Store/blob/master/app/src/main/java/com/zeapo/pwdstore/utils/PasswordRepository.java#L195

felixb:
done.

PR #317, opened 2017-07-16
Show extra content w/o copy password, fixes #288

### LINKED ISSUE #288: Extra content not shown until password copy
Disabling "Automatically Copy Password" causes the "Extra content:" box to remain hidden - no matter the setting of "Show Password". Manually pressing the copy button then makes the "Extra content:" fields visible.

The only time the R.id.crypto_extra_show text view is set to visible is in onPreExecute in DelayShow.
The only callchain I see to that is copyToClipBoard->setTimer->executeOnExecutor.

I believe that, at the very least, extra content should be set to visible in the PgpCallback onReturn if show_password is true. My personal preference would be for extra content to be visible even with the password hidden.

Thoughts on always making extra content visible?

felixb:
Show extra content even if password is not copied to clipboard.
Add toggle to preferences as well.

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/crypto/PgpHandler.java
+++ b/app/src/main/java/com/zeapo/pwdstore/crypto/PgpHandler.java
@@ -487,15 +487,18 @@ public class PgpHandler extends AppCompatActivity implements OpenPgpServiceConne
                     if (requestCode == REQUEST_CODE_DECRYPT_AND_VERIFY && os != null) {
                         try {
                             if (returnToCiphertextField) {
-                                boolean showPassword = settings.getBoolean("show_password", true);
+                                final boolean showPassword = settings.getBoolean("show_password", true);
+                                final boolean showExtraContent = settings.getBoolean("show_extra_content", true);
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/zeapo/pwdstore/crypto/PgpHandler.java
```diff
@@ -487,29 +487,31 @@ public void onReturn(Intent result) {
...
                         try {
                             if (returnToCiphertextField) {
-                                boolean showPassword = settings.getBoolean("show_password", true);
+                                final boolean showPassword = settings.getBoolean("show_password", true);
+                                final boolean showExtraContent = settings.getBoolean("show_extra_content", true);
                                 findViewById(R.id.crypto_container).setVisibility(View.VISIBLE);
 
                                 Typeface monoTypeface = Typeface.createFromAsset(getAssets(), "fonts/sourcecodepro.ttf");
-                                final String[] passContent = os.toString("UTF-8").split("\n");
+                                final String[] passContent = os.toString("UTF-8").split("\n", 2);
```
zeapo: nice ;)

gattazr:
Awesome !
Thanks for the feature :-)

Issue #192, opened 2016-06-29
Add more fields to autofill


jumper047:
I think it will be very useful to have easy bash to input login automatically. You can see example in pass firefox extension. It gets login name from 
"login: lapels" string, etc.

robinvd:
Does the current implementation do anything with the login or url data?

zeapo:
nothing much. we use the title of the password for autofill, but we do not decrypt the password (and extra content)

jumper047:
I also think, that it will be very useful. Some apps don't remember user name, or somebody can use more than one account on one device. 

27 сентября 2016 г. 12:53:37 GMT+03:00, Mohamed Zenadi <notifications@github.com> пишет:
>nothing much. we use the title of the password for autofill, but we do
>not decrypt the password (and extra content)
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/zeapo/Android-Password-Store/issues/192#issuecomment-249819620

-- 
Простите за краткость, создано в K-9 Mail.

PR #318, opened 2017-07-17
Parse username from password entry, refs #192

felixb:
This is a first step of letting APS handle usernames from password
entries. It shows the username in PgpHandler and allows to copy it to
the clipboard.


zeapo:
Great! Thanks for the PR, will try it later

zeapo:
Works pretty well :)

PR #320, opened 2017-07-25
Kotlin - part 1

zeapo:
This an initial work on a migration towards kotlin.

The targeted items in this part are:

- [x] User preference
- [x] Pgp Handler
  - [x] encryption
  - [x] decryption
  - [x] edit
  - [x] moving passwords (to be extracted form pgp handler as it has nothing to do here)
  - [x] get key ids

Next part will be Git


felixb:
Are you sure you really want to do this? IMHO, this will prevent people from contributing code easily. Aren't there enough other topics to work on?

app/src/main/java/com/zeapo/pwdstore/UserPreference.kt
```diff
@@ -334,9 +334,13 @@ class UserPreference : AppCompatActivity() {
...
         val sshKey = this.contentResolver.openInputStream(uri)
-        val privateKey = IOUtils.toByteArray(sshKey!!)
-        FileUtils.writeByteArrayToFile(File(filesDir.toString() + "/.ssh_key"), privateKey)
-        sshKey.close()
+        if (sshKey != null) {
+            val privateKey = IOUtils.toByteArray(sshKey)
+            FileUtils.writeByteArrayToFile(File(filesDir.toString() + "/.ssh_key"), privateKey)
+            sshKey.close()
+        } else {
+            Toast.makeText(this, "Unable to open the ssh private key, please check that the file exists", Toast.LENGTH_LONG).show()
```
felixb: I just externalized ALL the strings. Why adding new hardcoded strings into the code base?

zeapo: It's currently in WIP, no worry. I'm preparing also an update for French translation.

zeapo:
You're right on the contribution part. Is kotlin inaccessible for Java devs?

zeapo:
To be more precise why this choice. I started two years ago a fork of this app in Kotlin. However, at that time kotlin was not an officially supported langage. Now that [Google is officially supporting Kotlin for Android](https://developer.android.com/kotlin/index.html), it became obvious that more and more java developers will switch to Kotlin for Android development as it takes less time to write, it gives more null safety and is 100% compatible with java.

Moreover, many new features coming from Java 8, wont be available for minSdk < 24.

felixb:
Well, I'm not afraid of learning new languages. I just want to point out, that there are more people out there knowing java good enough to contribute some features or fix things.

zeapo:
@felixb Everything that has an `x` is feature complete. Could you check pls if something went wrong in the process?

Also, enjoy the shorter and clearer code compared to the previous PgpHandler ;)

felixb:
I can't test right now, because me phone is broken. :(
Did you think about adding some tests for your code?

zeapo:
@felixb Started adding some tests with espresso, the issue right now is that OpenKeychain has to be setup on the testing device for it to work. Do you have an idea how to set this up automatically?

I added the private key in the assets with a set of clear / encrypted passwords files too.

PR #321, opened 2017-07-25
Support pasting username with autofill, fixes #192

### LINKED ISSUE #192: Add more fields to autofill

felixb:
The workflow for pasting usernames is as follows:

1. Select password field
2. Select password store entry with username and paste it
3. Select any other editable field
4. Paste username

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/autofill/AutofillService.java
+++ b/app/src/main/java/com/zeapo/pwdstore/autofill/AutofillService.java
@@ -60,11 +60,10 @@ public class AutofillService extends AccessibilityService {
     private String webViewTitle = null;
     private String webViewURL = null;
     private PasswordEntry lastPassword;
-    private long lastPasswordDate;
+    private long lastPasswordMaxDate;
 
     final class Constants {
         static final String TAG = "Keychain";
-        static final long MAX_PASSWORD_ENTRY_LIFETIME = 45 * 1000;
     }
 
     public static AutofillService getInstance() {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/zeapo/pwdstore/autofill/AutofillService.java
```diff
@@ -58,16 +59,21 @@
...
     private String webViewTitle = null;
     private String webViewURL = null;
+    private PasswordEntry lastPassword;
+    private long lastPasswordDate;
 
-    public final class Constants {
-        public static final String TAG = "Keychain";
+    final class Constants {
+        static final String TAG = "Keychain";
+        static final long MAX_PASSWORD_ENTRY_LIFETIME = 45 * 1000;
```
zeapo: We could use the same timeout value as the copy/paste

zeapo:
Thanks for the PR.

In the workflow, could you add a notification (snack or toast, chose what you think is better) to inform the user that the selected password has a username in it and that the password was put in cache for `x` seconds so that he selects the username field to paste it.

app/src/main/java/com/zeapo/pwdstore/autofill/AutofillService.java
zeapo: It's `general_show_time`, we should rename it as it's used for visibility and clipboard timeouts.

felixb:
I added the toast in a 2nd commit for easier review. Feel free to squash the commits when merging.

zeapo:
Works perfectly, thanks.

PR #327, opened 2017-07-31
Update Russian translation

bshkrv:
It's been a while since I contributed translation. Here is the updated version.


zeapo:
thanks for the update

Issue #330, opened 2017-08-01
Show password age 


chrboe:
It would be nice to be able to display the "age" of a password (i.e. days since last change or something). Old passwords (older than 30, 60, 90,... days) could additionally be highlighted in red or a similar alarming color.

zeapo:
Good idea, we can use git commits to know when the file was last updated.

chrboe:
Exactly. It's actually pretty easy, I've written something similar in go before: https://github.com/chrboe/pass-age

chrboe:
And the same thing as a bash one-liner, for inspiration :)
`git ls-files *[^.gpg-id] | while read f; do echo -e "$(git log -1 --format="%ai %ar" -- $f) \t $f"; done | sort -r | cut -d' ' -f4-`

PR #339, opened 2017-09-10
show age for passwords in detail view

chrboe:
Implements #330.
This fetches the latest commit where the respective password file was
changed from the current HEAD and outputs the relative time since
the last change on the decrypt page.

NOTE: translations are incomplete because I only speak so many languages :)
If someone hits me up with other translations I'll be happy to include them
in this PR.

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
+++ b/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
@@ -42,7 +42,9 @@ import com.zeapo.pwdstore.utils.PasswordRepository;
 
 import org.apache.commons.io.FileUtils;
 import org.eclipse.jgit.api.Git;
+import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.lib.Repository;
+import org.eclipse.jgit.revwalk.RevCommit;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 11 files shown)


zeapo:
Thanks for the PR. However I do not want to introduce any "git" logic in the decryption / encryption process. Especially if we want (and will) support non-git stores.

chrboe:
That's understandable. So do you suggest just getting the last changed date from the actual file if it's a non-git store then?

zeapo:
In the case of git repository we should keep the commit technique. However I'd like it to be "outside" PgpActivity as it should be agnostic to what store it's in. Especially that in a near future it won't open any file, it will just receive bytearray and decrypt/encrypt them :)

Maybe you could move the logic in PasswordStore, and pass the age (in days) as an intent data to PgpActivity?

chrboe:
Yes, that would seem like a sensible enough approach. I'll see if I can find some time to take a stab at this today

zeapo:
Thanks for the update, I will try it later today :)

zeapo:
Sorry for the delay @chrboe , I will take care of your PR with the two others ASAP :)

zeapo:

zeapo:
@chrboe could you please rebase so that I merge this :)

chrboe:
Sure thing, I'm on it :) Thanks for taking the time to review this!

chrboe:
Phew, at last. I had a few issues along the way, but it should be all fine now.

Issue #326, opened 2017-07-31
pass-otp support


juicetin:
Hi,

Would support for the OTP format in pass-otp (e.g. `otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example
`) happen to already be in the works? 
https://github.com/tadfisher/pass-otp#readme

The underlying logic probably wouldn't be hard to implement, though to have the feature integrated nicely (vs. manual file creation/typing out of the parameters in the above format) may require a little bit of extra work with having the 'Add New Password' screen having something like a dropdown near the top to be able to choose between 'password' and 'OTP' or something along those lines.

zeapo:
Absolutely!

Thanks for the idea.

maximbaz:
I wonder if it would make sense to approach this iteratively, first implement generating tokens from already configured OTP accounts, and then separately implement configuring OTP for new accounts. I think it's quite safe to assume that people who know about `pass-otp` have all their accounts already configured, so generating tokens from existing configuration should be a much more requested feature than configuring new accounts. Also, hopefully should be simpler to implement! 🙂

c4software:
Nice idea. Would be really great ! I have it on my desktop, it's would be really great to see the support of otp on the mobile too.

polyzen:
Support included since Jan.

erayd:
@polyzen Not on the Play store it's not. That hasn't been updated since November 19th last year.

Any chance whoever is responsible for the relevant Play account could roll a new release there?

polyzen:
> @polyzen Not on the Play store it's not. That hasn't been updated since November 19th last year.

If you use F-Droid, that's up-to-date.

erayd:
> If you use F-Droid, that's up-to-date.

Thanks - That's good to know :-). The Play store app still needs updating though.

AluminumTank:
Is HOTP support still being worked on?

wiktor-k:
Nope, it you check the Readme you'll see that the project is currently unmaintained.

PR #349, opened 2017-11-06
Display TOTP code if password contains OTP secret

wiktor-k:
This is a work in progress implementation of issue https://github.com/zeapo/Android-Password-Store/issues/326.

If a password is a specially formatted TOTP URI then a number is calculated from the secret and displayed below the username. If the password is not a TOTP URI then the entire line is hidden (similarly to username/login).

It looks like this:

totp

Open questions:

- [x] Which base32 library to use, I attached something that uses BSD licence, there is a snippet from Google with Apache licence (https://github.com/aerogear/aerogear-otp-java/blob/master/src/main/java/org/jboss/aerogear/security/otp/api/Base32.java) or we could use Apache Codec but that's importing entire lib just for one simple class.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -68,6 +68,7 @@ dependencies {
     }
     compile 'com.jcraft:jsch:0.1.54'
     compile group: 'commons-io', name: 'commons-io', version: '2.4'
+    compile group: 'commons-codec', name: 'commons-codec', version: '1.11'
     compile 'com.jayway.android.robotium:robotium-solo:5.3.1'
     compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlin_version"
     compile 'com.android.support.constraint:constraint-layout:1.0.2'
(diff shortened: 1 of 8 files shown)


zeapo:
Thank you for the PR. 

Can you use the same logic behind the "username" field to detect the OTP? It would be better to be able to store the username, OPT as an extra rather than the password itself.

I'll go for using apache-codec, the apk size wont increase much (we have to see ;)

wiktor-k:
Hmm... I placed it in the password field to maintain compatibility with pass-otp (see #326) but on a second glance it seems the extension scans all lines looking for [strings starting with `otpauth://`](https://github.com/tadfisher/pass-otp/blob/develop/otp.bash#L234-L240).

I'm not using pass-otp but I'd want to maintain compatibility with it (no need to invent another format).

Would it be OK to scan all lines for `otpauth://`? It's not a line in the usual pass' `key: value` format...

zeapo:
Absolutely!
You can do it so that it scans all the file, not just extras. That way you cover both cases (first line, or other lines).

wiktor-k:
Oookay, now it should look a little bit better...

zeapo:
nothing to be done, I just didn't have time to work on APS :(

zeapo:
@wongma7 do you have few moments to test this?

wiktor-k:
@zeapo got it, thanks!

zeapo:

wiktor-k:
No worries, I just don't want this to be in forever-open state.

lecler-i:
This doen't work for me if the `otpauth://` is not on the first line.
It shows the line into the `Extra content` without taking it into consideration

PR #354, opened 2017-11-19
Remove autofill-accessibility

zeapo:
For the moment this is just a comment of autofill service and any access to it.

This will be the build for Play Store untill we implement the keyboard and android-8 autofill


zeapo:
No longer relevent. We'll have to deal with any future request from Google when that comes ^_^

Issue #355, opened 2017-11-19
Update icon


shgysk8zer0:
This is a sub-issue for #352 
Original SVGs may be found [in this Gist](https://gist.github.com/shgysk8zer0/9436084ca74adef73306f06fc2b697ed)

The current new icon was intended to not look too foreign when updated. Here are some suggestions on iterations of that icon.

![inverted](https://user-images.githubusercontent.com/1627459/32995985-791384d4-cd31-11e7-8d05-f35d1aa1b1a8.png)

![alt](https://user-images.githubusercontent.com/1627459/32995989-86c9ba94-cd31-11e7-990f-bc1191cec9fb.png)

![no-lock](https://user-images.githubusercontent.com/1627459/32995992-8b56221e-cd31-11e7-8f45-8ce6d58ba29f.png)

I prefer the third. That seems to be the popular (thus far, unanimous) opinion as well.

I can make changes to any of these, but wanted to share the mindset and reasoning behind choice of color, rotation, shapes.
- As you probably recognize, it is based on the Git logo
- The descending portion replaces on of the branches of the Git logo with a common symbol for passwords
- When viewed as a whole, it also makes the shape of a keyhole (would not really imply the same if rotated)
- The orange color in 2 & 3 is also taken form the Git logo
- The blue on the second is complimentary to the orange

I'm open to suggestions. I'll try to incorporate anything that is popular, I think is a good idea, or comes from @zeapo .

erayd:
I very much like the third, but would like to suggest rotating it -90 degrees ([SVG source](https://gist.github.com/erayd/e360dd83afd76e0184841f37dc24db44)). IMO this is more obviously referencing a typed password field with version history.

![android-password-store-no-lock-rotated--90](https://user-images.githubusercontent.com/121486/32996427-7ec8c77a-cde7-11e7-81cc-c58a2c48257d.png)

shgysk8zer0:
I can see that. I'm mixed though because it loses the shape of a key hole and makes the relation to Git less obvious.

I'll try to think of a way of getting the best of both. Else, I leave it up to the owner.

PR #356, opened 2017-11-20
Use adaptive icon

dasJ:
This adds support for an adaptive icon, as seen in Android Oreo.

As the icon was not available as svg, I recreated it from scratch and also commited it into the `contrib/icons` directory.

adaptive icons


zeapo:
Hi, thank you for your work. 
Have you checked the #355 ? The icon you based on was flagged by Play Store as a copyright infringement (which is not... and the appeal didn't go through even though the license permits it)

If you can improve upon the #355 that would be nice.

dasJ:
Yes, sorry. When I created and opened the PR, I haven't found #355 yet ;)

PR #362, opened 2017-12-27
Password should be invisible to the accessibility services

mohammadnaseri:
Due to recent attacks, malicious apps that are using the accessibility service, can capture all user inputs. In this case, password, should be ignored for the accessibility service, so such attacks can not be happened in this app.


zeapo:
Thanks for the nice PR

PR #367, opened 2018-01-04
Fix totp elements layout

balasankarc:
TOTP element was overlapping `username` label. This PR fixes it
  
Before (see how TOTP is above Username):
before

After:
after


zeapo:
Thanks for the PR.
What would happen if there's no username? Wouldn't that leave a space above TOTP?

balasankarc:
@zeapo True. It will. I had thought about it. Unfortunately that bug is a pre-existing one, one that predates introduction of totp logic. If an entry has extra content and no username, there will be a blank space above extra content.

I will think about how we can fix it quickly, but for now I just focused on getting totp usable and didn't think about the aesthetics much.

balasankarc:
@zeapo I think 8c270f5 fixes it.

zeapo:
looks good, thanks

PR #368, opened 2018-01-05
Update README.md

ggarek:
I suggest to make picture width 320px. Currently it is very big and does not even fit the screen.
And it seems, `style` does not work for some reason.

- [x] set usage gif width to 320px
- [x] use `width` property instead of `style="width: ..."`


zeapo:
thanks for the fix

PR #407, opened 2018-05-10
fix zh_rCN string translate.

cpdyj:
Thanks.


zeapo:
thanks for the update

PR #413, opened 2018-06-05
Display HOTP code if password contains HOTP secret, unify HOTP and TOTP code

AluminumTank:
This is a follow-up to PR #349 and issue #326.

This should work the same way as TOTP support has worked in Android-password-Store, with the only difference being that the `otpauth://` string must be an hotp string instead.

EDIT: I should mention that, since the counter in the `otpauth://hotp/...` string is ~~automatically incremented every time the entry is opened~~ incremented every time the user requests, this can lead to issues with conflicting git commit histories on different machines.


zeapo:
Hey, thanks for the PR.

The fact that we have to auto-commit every time the user decrypt the password is an issue by itself :/
We already have issues with git conflicts :(

AluminumTank:
I'm not sure I'm familiar with the git conflict issues that Android Password Store deals with. Unless you're referring to user error that leads to git conflicts.

zeapo:
Not mandatory a user error. If the user changes his password on his terminal, and the counter is updated on his device. this will result in a conflict.

zeapo:
unless we enforce auto-sync for every change...

zeapo:
Could you add a warning that it's going to auto-commit?
It will be weird when users notice that their files are changing without them knowing.

zeapo:
I'd go for as soon as we detect it. Tell them that we're about to update
the file as it contains an htop secret. If he agrees, go ahead, otherwise
just show the content. Plus a "remember my choice" or something like that.

Sorry that will add a little bit of work.



Le mar. 28 août 2018 à 19:24, Joel Beckmeyer <notifications@github.com> a
écrit :

erayd:
> I'd go for as soon as we detect it. Tell them that we're about to update
> the file as it contains an htop secret. If he agrees, go ahead, otherwise
> just show the content. Plus a "remember my choice" or something like that.

I think this is the right approach.

For what it's worth, I personally **never** want the git repo updated from my phone, and it uses a read-only key to access the repo. If something decided to just update things without asking me, this would be extremely annoying.

zeapo:
Thanks @AluminumTank.

Didn't try it, but looks good to me. I'll let @zidhuss test this and make his first merge :)

Issue #419, opened 2018-07-25
"Clear clipboard 20 times" not working.


alexlab2017:
First of all: this is a GREAT app!

I have a Samsung Galaxy S9+ with version 1.2.0.75 of Password Store.
My device is brand new with stock UI and Android (Samsung Experience 9.0, Android 8.0.0) and has clipboard history so I enabled the option "Clear clipboard 20 times".
I watched the clipboard after having copied a password, what happens after 45 secs is that the clipboard start being filled with progressive numbers from 0 to 18 (so 19 entries in the clipboard) and the oldest entry (20th entry) still is the copied password.

I'm not an Android developer but I think this might be fixed adding one more copy and this will not have any negative impact because in the worst case will be just redundant; anyway this is just a suggestion.

To do a double check I tested this option with my old phone (Huaewi Ascend Mate 7) that doesn't feature clipboard history and what happens if I paste after 45 secs is that I paste "18".

The other thing I noticed that in the description of the option is written "Store nonsense in the clipboard 20 times instead of just once. Useful on Samsung phones that feature clipboard history." but actually what is stored isn't nonsense but progressive integers (0-18).

Let me know if I can provide any additional data.

Thanks for the attention,
Alessandro

PR #423, opened 2018-09-02
Clear clipboard 20 times fix

alexlab2017:
As described in issue #419 there was a bug when the preference meant to clear clipboard history, specially on Samsung phones, was selected, actually clearing the clipboard one time less than the needed amount (19 times vs 20 times).


zeapo:
Thanks!

alexlab2017:
As already said: GREAT app.
Happy that I could be helpful.

PR #425, opened 2018-09-20
Clean paths

cortex:
This changes display of paths in password listings, so only necessary ones are shown. Slash is stripped from the end of folders (they already have an icon), and from start and end of paths. I think this gives the app a cleaner appearance. 

I am not entirely sure about the item layout, what I did here looks better imo, but it could be easier to do if converted to a ConstraintLayout, but I don't know what is the preference in this project.


zidhuss:
Looks great, thanks for the PR!

Issue #416, opened 2018-07-10
Does not copy password anymore after loading several entries


AuspeXeu:
I am using version `1.2.0.73` when selecting the first entry, the password gets copied to the clipboard. Same for second and third (I use the back button to go back to the main directory of entries after one entry is copied to the clipboard). The fourth time I click on an entry, it won't show the password nor copy it to the clipboard.

PR #431, opened 2018-10-02
Cancel running DelayShow async tasks (#416)

setine:
These tasks were filling up the threadpool slots and leading
to delays in executing further AsyncTasks after a while.


zeapo:
Thanks for the pr!

setine:
Np, was bugging me, too :) Thank you for all your work on this app!

Issue #302, opened 2017-05-28
Add full path to commit messages 


jayme-github:
Currently just the basename of the password store file is added to the git commit messages:
```
    [ANDROID PwdStore] Add  anotheruser  from store.

    [ANDROID PwdStore] Edit  theuser
```

As this leads to duplicate messages for multiple identical usernames in different directories it would be nice if the commit messages would contain the full path like the ones from "pass" do:
```
    Add given password for tests/Linux/theuser to store
```

zidhuss:
Added in #455.

PR #455, opened 2018-11-30
Commit messages

zidhuss:
Making the commit messages match those of the command line app. Also addresses #302 

What's left to do:
- [x] Update translated strings to use the [format strings](https://developer.android.com/guide/topics/resources/string-resource#formatting-strings).
- [x] Use the move/rename string in commit message


zidhuss:
Commit message now look like this:

Adding passwords:
```
Add password for zidhuss using android password store.
```

Moving/renaming passwords:
```
Rename zidhuss to github/zidhuss.
```

Deleting passwords:
```
Remove github/zidhuss from store.
```

PR #460, opened 2018-12-14
Clipboard now is cleared after manual copy and Spanish translation is added

reroman:
The password now is cleared in both cases, when is manually and automatically copied to clipboard. Also if the option is deactivated, only show the message "The Password is copied to clipboard". This fix the issue #374. Also, Spanish translation was added.

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
+++ b/app/src/main/java/com/zeapo/pwdstore/PasswordStore.java
@@ -516,8 +516,8 @@ public class PasswordStore extends AppCompatActivity {
 
         if (settings.getStringSet("openpgp_key_ids_set", new HashSet<String>()).isEmpty()) {
             new AlertDialog.Builder(this)
-                    .setTitle("OpenPGP key not selected")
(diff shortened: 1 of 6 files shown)


zidhuss:
@reroman thanks for the pull request!

Could you update your Spanish translations for the git commit messages to make use of format strings? As in: https://github.com/zeapo/Android-Password-Store/blob/master/app/src/main/res/values/strings.xml#L25.

zidhuss:
Is this also related to #186?

reroman:
> @reroman thanks for the pull request!
> 
> Could you update your Spanish translations for the git commit messages to make use of format strings? As in: https://github.com/zeapo/Android-Password-Store/blob/master/app/src/main/res/values/strings.xml#L25.

Ok, no problem.

zidhuss:
Hey @reroman, again thanks for your contribution. I thought it would be best to split this pull request as it includes two separate changes, i.e. clearing the clipboard and the Spanish translation.

I went and added the translations to the master branch (e58a5e7). Could you rebase and open a new pull request with the clipboard clearing code (6b8ac9a)?

Issue #461, opened 2018-12-26
The CheckBoxPreference for clear_after_copy preference is unnecessary


reroman:
I was reviewing the code to fix some issues related to the cleaning of the clipboard and I found that the timer to clean the clipboard is not started if the `general_show_time` value is zero. Currently I'd have to validate two parameters that could oppose each other. For example, if `general_show_time` is non zero, but `clear_after_copy` is not enabled, then the value of the first one is non sense. In the other way, if `general_show_time` is zero, the value for `clear_after_copy` doesn't matter.

My conclusion is (according to the summary of the preference) to control the setting only with `general_show_time` and add a note for value of zero, something like _"Set the time you want the password to be in clipboard before cleared. 0 means auto clear disabled"_. I think this is the easiest solution for over validations.

Although I'm not very sure if `general_show_time` was designed to control the time of the clipboard or just the time of the showed activity (it could be both, but this only makes sense if autocopy is always enabled, unless there was two timers). If it's to control the activity, then there's a bug because the timer is only started when the password is copied and not when is decrypted.

PR #464, opened 2019-01-02
Clipboard now is cleared with manual copy

reroman:
The clipboard is automatically cleared only if the password was copied (manual or automatically) and if `general_show_time` is non zero. Also the checkbox for `clear_after_copy` preference depends on the value of the `general_show_time` to avoid the oppose of each other.

This fixes the issue #374 and #186. But I still it's redundant to have the two options, the checkbox and the time for showing the password (as I describe in #461).


zeapo:
Thanks for the PR. Could you address the comments above and mark them as Resolved, plz? 

I'll merge then :)

reroman:
> Thanks for the PR. Could you address the comments above and mark them as Resolved, plz?
> 
> I'll merge then :)

Done!

franok:
really looking forward to see this important fix merged & released on F-Droid 😊 
I ❤️ pass/Android-Password-Store

PR #411, opened 2018-05-25
WIP: Autofill for Android > Oreo (API 26)

l0rn:
This is my attempt to make use of the new autofill api

PR #471, opened 2019-01-06
Enable parallel installation of release and debug builds

dmo60:
Hi! Thank you for this great app! 

I wanted to play around with the code and ultimately investigating if I could contribute to the Autofill implementation from #411. But the first thing I noticed is that it's currently not easily possible to have both a release and a debug build installed at the same time. 

This PR addresses this by specifying an `applicationIdSuffix` for the `debug` build type. With that, the debug build will have a different application id (`com.zeapo.pwdstore.debug`) and thus will not collide with the release app when installed on the same device.

This would allow for a "dev"/"testing" environment, where one could e.g. setup a different password store for development purposes. At the same time, the release build would still be available for day-to-day usage.

What one would probably want in this case is a different icon for the debug app to visually differentiate between the two apps in the application launcher. This is where I would need some help, since I neither have any skills in manipulating Vector images, nor do I have access to the source SVG file that was used to generate the launcher icons (it would be nice if this svg source could be added to the repository). Usually a simple "DEBUG" text badge would do the job.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -5,6 +5,7 @@ apply plugin: 'eclipse'
 
 android {
     compileSdkVersion 28
+
     defaultConfig {
         applicationId "com.zeapo.pwdstore"
         minSdkVersion 21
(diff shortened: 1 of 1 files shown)


zeapo:
Nice :)

Thanks for the PR

zidhuss:
Hey @dmo60, could we possibly change the background colour for the debug release?

https://github.com/zeapo/Android-Password-Store/blob/master/app/src/main/res/values/colors.xml#L18

zeapo:
@zidhuss oh sorry, you're right. I didn't think about that. The PR is merged, can you make the change to color you like?

dmo60:
> Hey @dmo60, could we possibly change the background colour for the debug release?

Yep, that's what I actually had in mind as well. I will create a PR for that. Still I think it would be nice if we would indicate the "Debug" in the icon itself as well.

dmo60:
see #472

PR #474, opened 2019-01-10
Add option for default directory on open

### LINKED ISSUE #446: Doesn't show hidden folders anymore
I use APS directly with my dotfiles repo. Starting with 1.3.x it doesn't see .password-store folder, nor does it see any other dot folder.

etu:
This could work around the issue with people having their
.password-store as a directory in the repository. Because they can
specify that directory to be the default directory and pop in there
automatically.

Also this would make a nicer experience for people with other things
in the same repository since they will only see relevant files and
directories.

Known problems:
Changing the setting doesn't do anything until the app is restarted.

But bear with me here. I have never touched Java, Kotlin, Android or Android Studio before in my life.

I believe that this fixes #446

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/utils/PasswordRepository.java
+++ b/app/src/main/java/com/zeapo/pwdstore/utils/PasswordRepository.java
@@ -118,13 +118,15 @@ public class PasswordRepository {
         File dir = null;
         SharedPreferences settings = PreferenceManager.getDefaultSharedPreferences(context.getApplicationContext());
 
+        String defaultDirectory = settings.getString("default_password_directory", "");
+
         if (settings.getBoolean("git_external", false)) {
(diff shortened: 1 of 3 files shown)


zeapo:
Thanks for the PR.

However, a big issue there. The `getRepositoryDirectory` method has to match the `.git` directory. At least for the moment. Which means that using your changes, you wont be able to commit any file.

I know that the current structure is bad... really bad, we have to rewrite it completely. 

I prefer if, for the moment, we go for the option to show the hidden files.

etu:
@zeapo I have tested this on my actual device, with a git-repo. And it works to read passwords. I didn't think of creating one because it's something that I never really done on my phone.

But at least I have a build that works for me until it's solved from upstream so I'm happy with it anyways :-D

Issue #395, opened 2018-04-10
Cannot create numerals only passwords


reroman:
When I only select Numerals in the Generate Password dialog, the password generated always contains letters.

zidhuss:
With all the flags unset, it defaults to generating a password with only lowercase characters. When you're selecting to include `Numerals`, you're allowing numerals to be generated along with the lowercase characters.

However we probably _do_ want the ability to generate numeral only passwords.

zeapo:
To generate numerals only, we should add a checkbox "include letters". When unchecked it would remove [a-zA-z] from the generator.

FabianHenneke:
Resolved by #478.

PR #478, opened 2019-01-15
Allow to create passwords without lowercase letters.

reroman:
This now allows to create passwords without lowercase letters, for example when a pin number is required. It solves problems like #395.

I added a 'L' to the current flags for this purpose.


zeapo:
nice idea! Thanks

Issue #71, opened 2015-03-17
Use GPG Key for SSH


the-kenny:
One can use GPG subkeys as SSH keys. I store/generate my keys on a Yubikey which allows using them under Android.

It would be really nice if android-password-store could use the keys on the smartcard via OpenKeychain so I don't have to store the ssh-key on the device.

zeapo:
That would be great indeed.

zeapo:
From my investigations, we will have to write some code for both OpenKeychain app and openpgp-api-lib to support that. Neither of them offer an intent (or api for the later) to export a key or sub-key.

regarding the ssh-key import, you can delete the key once you imported it. we make a copy inside the data folder to avoid that other apps access it without permission (root necessary)

shtrom:
:+1:

jfjallid:
Any progress on this? This feature would be really useful!

zeapo:
This will have to wait for https://github.com/open-keychain/open-keychain/issues/908

jcgruenhage:
Afaict the open-keychain part has been done, and PRs have been submitted to ConnectBot already.

saro:
Apparently https://github.com/open-keychain/termbot is a fork of connectbot with the new Authentication API. Isn't clear from https://github.com/connectbot/connectbot/pull/567 if this feature will be included in the main ConnectBot project.

languitar:
Can someone explain why connect bot is in the loop here? Or is this just a sidenote that others are already using the API?

jcgruenhage:
@languitar just a sidenote

colemickens:
@alex005005 you'd mentioned working on this, any progress? I switched my entire life over to `gopass` before realizing that this isn't implemented (I mistook the existing OpenKeychain support) so I'm keen on trying to get this working. I don't know much about Android, but I know enough Java to probably copy code from TermBot to this app.

colemickens:
Heh, looks like more code than I expected, go figure. In case someone more proficient in Android is interested, it looks like the TermBot authors are still working on upstreaming to ConnectBot  so it's actually quite easy to see the diff to add support for OpenKeychain here: https://github.com/connectbot/connectbot/pull/567

zidhuss:
Added in #486.

PR #486, opened 2019-02-25
Authentication using OpenKeystore SSH API

vexofp:
Implementation for #71 . It seems to work well with my Yubikey 5 over NFC.

Integrating the OpenKeychain API's model (lots of PendingIntents) with the existing code (mostly modal dialog boxes) was a bit interesting. The result is not quite as elegant as I would like, but manages to avoid any intrusive refactoring of the existing authentication methods.


razic:
@vexofp very nice

app/src/main/java/com/zeapo/pwdstore/git/GitActivity.java
```diff
@@ -676,23 +681,62 @@ protected void onActivityResult(int requestCode, int resultCode,
...
+            // IdentityBuilder attempt to extract some updated state from the intent before
+            // trying to re-launch the operation.
+            if (identityBuilder != null) {
+                identityBuilder.consume(data);
+            }
+            launchGitOperation(requestCode);
         }
     }
 
+    @Override
```
MSF-Jarvis: Useless override? You're not doing anything here

zeapo:
@MSF-Jarvis we have only few tests ;(

I can start looking into this. For the moment, this needs manual testing

zeapo:
Tested manually. It works fine. Openkeychain is surprisingly simple to use for these kind of use cases :)

MSF-Jarvis:
> Tested manually. It works fine. Openkeychain is surprisingly simple to use for these kind of use cases :)

Sweet, I'll rebase and merge this.

zeapo:
@vexofp I get asked to select the key on every call to jgit. Is it possible to make that selection when the user opt for OpenKeychain authentication?

Also, can you add an item in the FAQ on how to create an Authentication subkey in OpenKeychain?

razic:
I tested this feature manually as well. It works great. Thank you very much @vexofp. To me personally, this was the feature I was missing the absolute most.

@zeapo's suggestion with regard to saving the preferred key such that you are not prompted on every git operation would definitely be a good improvement to have.

Thanks again! Looking forward to seeing this merged.

bkchr:
Works for me as well :) Would be nice to have this into official releases!

app/src/main/java/com/zeapo/pwdstore/git/GitActivity.java
vexofp: Yep, this was left-over from debugging.

vexofp:
I believe I have addressed the last few review comments and added a section to the [wiki](https://github.com/zeapo/Android-Password-Store/wiki/Generate-SSH-Key#from-openkeychain-coming-soon) on generating authentication keys in OpenKeychain, as requested by @zeapo.

I can try to work on saving the chosen key ID when I have some time, but that might be better as a separate PR.

bkchr:
Nice :) Hopefully this gets merged then :)

zeapo:
I'll test this tomorrow and then merge it.

Le mar. 2 avr. 2019 à 11:22, Bastian Köcher <notifications@github.com> a
écrit :

therealpxc:
What an exciting feature!

zeapo:
thanks :)

Issue #465, opened 2019-01-02
Clipboard history is not cleared 20 times


tomasz-oponowicz:
### Steps

1. Given the clipboard history is cleared
1. Given the "Clear clipboard 20 times" is enabled in Password Store
1. Given the "Automatically Copy Password" is enabled in Password Store
1. Given the "Automatically clear the clipboard adter..." is enabled in Password Store
1. Open password store
1. Open an entry. A password will be copied automatically.
1. Wait till the clipboard clearing is triggered
1. Open a Samsung keyboard (for example use a text field in any app)
1. Go to _Clipboard history_

### Expected

The clipboard history doesn't contain the copied password.

### Actual

The clipboard history contains following entries:

`18, 17, ... 0, <copied password>`

### Environment

Android: 8.0
Version: 1.3.2

ashfaqfarooqui:
I have the same issue on my Samsung phone.

zeapo:
fixed in  #487

PR #487, opened 2019-02-26
Fix clear clipboard 20 times history (#465)

rk0n:
I tested it on a Samsung Galaxy S9+ and the clipboard history is now clearing properly.


zeapo:
thanks for finding that off-by one :)

PR #701, opened 2020-04-14
Always refresh password list when navigating back

FabianHenneke:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Ends the practice of caching contents of folders on the navigation stack.

Caching and restoring directory contents introduces a lot of logic that can lead to bugs and goes against the spirit of the LiveData approach. Since file listing now happens on the IO thread and the scroll position is restored exactly, it is not even clear we need this kind of caching.

I verified that navigation still works as it previously did (and better in at least one edge case).

## :pencil: Checklist

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/SearchableRepositoryViewModel.kt
+++ b/app/src/main/java/com/zeapo/pwdstore/SearchableRepositoryViewModel.kt
@@ -43,7 +43,6 @@ import kotlinx.coroutines.flow.emptyFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.mapLatest
-import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.flow.toList
 import kotlinx.coroutines.yield
 import me.zhanghai.android.fastscroll.PopupTextProvider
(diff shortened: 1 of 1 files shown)


msfjarvis:
Do you wanna get rid of the caching? What would that entail for the end user?

FabianHenneke:
> Do you wanna get rid of the caching? What would that entail for the end user?

Since navigating back does not trigger a search, but rather a simple file listing without filtering, I think that this should not cause a noticeable performance hit now that we are doing this on the IO thread. Furthermore, we have only recently begun to restore scroll positions approximately and the new implementation even does this exactly, which should more than make up for any lags.

I would vote for disabling it for the release. If there should be any real-world performance impact by this, we could revisit and rethink it (and all the edge cases).

msfjarvis:
> > Do you wanna get rid of the caching? What would that entail for the end user?
> 
> Since navigating back does not trigger a search, but rather a simple file listing without filtering, I think that this should not cause a noticeable performance hit now that we are doing this on the IO thread. Furthermore, we have only recently begun to restore scroll positions approximately and the new implementation even does this exactly, which should more than make up for any lags.

Agreed.

> I would vote for disabling it for the release. If there should be any real-world performance impact by this, we could revisit and rethink it (and all the edge cases).

Wanna update this PR for that? I'd love for a broader feedback loop on this.

FabianHenneke:
> Wanna update this PR for that? I'd love for a broader feedback loop on this.

I has already been in there from the start ;-)

msfjarvis:
> > Wanna update this PR for that? I'd love for a broader feedback loop on this.
> 
> I has already been in there from the start ;-)


whaaaaaaa

PR #743, opened 2020-04-24
Fix search if external root dir starts with a dot

### LINKED ISSUE #740: [BUG] cannot search in root folder
**Describe the bug**
I have two folders in my password store root, the search won't work when I'm at root, only works if I go into one of them. 

**To Reproduce**
1. Go to home page
2. Click on search
3. type anything

**Expected behavior**
display search results from both folders

**Screenshots**
image
image
image

 - Device: Google Pixel 3
 - OS: stock ROM Android version 10.0
 - App version: 1.7.1

**Additional context**
The Recursive filtering is checked in Settings.

Hazegard:
- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description
This PR aims to not filter out the root folder.

This PR fixes a small bug I encountered.
My repository is located at /storage/emulated/0/.password-store.

However, in the autofill popup, since that my root folder is hidden, it is filtered out.
So the autofill menu was empty.

Edit: Fixes #740.

I checked that the autofill menu now shows entries

## :pencil: Checklist

### DIFF
--- a/app/src/main/java/com/zeapo/pwdstore/SearchableRepositoryViewModel.kt
+++ b/app/src/main/java/com/zeapo/pwdstore/SearchableRepositoryViewModel.kt
@@ -266,7 +266,7 @@ class SearchableRepositoryViewModel(application: Application) : AndroidViewModel
 
     private fun listFilesRecursively(dir: File): Flow<File> {
         return dir
-            .walkTopDown().onEnter { file -> shouldTake(file) }
+            .walkTopDown().onEnter { file -> if (file == dir) true else shouldTake(file) }
             .asFlow()
             // Skip the root directory
             .drop(1)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/zeapo/pwdstore/SearchableRepositoryViewModel.kt
```diff
@@ -266,7 +266,7 @@ class SearchableRepositoryViewModel(application: Application) : AndroidViewModel
 
     private fun listFilesRecursively(dir: File): Flow<File> {
         return dir
-            .walkTopDown().onEnter { file -> shouldTake(file) }
+            .walkTopDown().onEnter { file -> if (file == dir) true else shouldTake(file) }
```
FabianHenneke, line 269:
```suggestion
            // Take top directory even if it is hidden.
            .walkTopDown().onEnter { file -> file == dir || shouldTake(file) }
```

FabianHenneke: @Hazegard I find it a bit clearer this way, what do you think?

Hazegard: Thank you for your suggestion, I agree it's better!

PR #905, opened 2020-07-01
Improve and refactor Autofill heuristics

fmeum:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## :scroll: Description

* Add support for `AUTOFILL_HINT_NEW_PASSWORD` and
  `AUTOFILL_HINT_NEW_USERNAME`. This allows apps to trigger a
  `ClassifiedScenario` with only a generate password action and is the
  analogue of the W3C new-password hint for websites.
* Do not consider HTML password fields without hints to be certain
  password fields (they could contain e.g. bank account numbers,
  API secrets,...).
* Reduce OTP field false positives by excluding the term "postal" as well
  as fields that match the "code" heuristic term but have HTML maxLength
  less than 6 or larger than 8.
* Add German heuristic term "einmal" ("one-time") for OTP fields
* Also exclude fields based on their HTML name (e.g. for terms such as
  "search").
* Extract fieldId, hint and htmlName matches into an extension property.
* Reduce warnings and remove unnecessary suppression annotations.

Only lightly, but these kind of changes can only really be tested by snapshot users.

## :pencil: Checklist


msfjarvis (approved):
Changes look sane to me, I'm able to fill into my usual sites.

PR #960, opened 2020-07-23
Fix OpenPgpApi extra type confusion

fmeum:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## :scroll: Description

`Array<Long>` and `LongArray` are distinct types and the former is being passed to Java as a `Serializable` rather than a `Long[]`. With this change, encrypting to fingerprints and long key IDs of master keys now works.

Note that providing the key ID of an encryption subkey does not seem to be supported by OpenKeychain.

## :pencil: Checklist


msfjarvis (approved):
Allow me a moment while I tear my hair out at this idiocy.

PR #1056, opened 2020-08-26
Add Activity to view the Git commit log 

### LINKED ISSUE #284: Add ability to view git commit log and SHA's

Nosweh:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## :scroll: Description
The GitLogActivity is accessible via the action bar menu ("View commit log") if the password repo is a Git repository. The Activity displays basic information about the commits in the repository using a RecyclerView, after internally performing a git-log command. This information is

1. the short commit hash (first 8 characters)
2. the commit's short message
3. the commit date and time.

(Also see the screenshots.)

I'm not quite sure if we should also display the author of the commits because I suppose for most users they will be always the same. Or is it a common use case to have multiple authors contribute to the same pass repo? It would not be a problem to add it, though. The (newly added) intermediate data class GitCommit does contain the author's name already; I just decided to remove it from the UI for now, for simplicity. What do you all think?

See open issue  #284.

To test the UI and usage of a 'casual' user, I opened a password Git repo in APS with a couple of commits and then opened the new Activity to view the "Commit log". I performed this test on three devices:
- an AVD phone with 1080x1920 resolution at 420 ppi running Android 10,
- an AVD tablet with 2560x1800 resolution at xhdpi running Android 10, and
- a physical phone with 1080x2280 resolution at 432 ppi running Android 9.

To test the performance of acquiring the commit metadata, I opened a password Git repo in APS with 1,000 commits and then viewed the commit log. I performed this test on a physical, mid-range 2018 phone running Android 9 (+ on an AVD, to be precise). I noticed no noticeably longer delay in the user experience compared to performing other user actions in the app.

## :pencil: Checklist

## :crystal_ball: Next steps
We could make the items in this new Git commit log interactive to show more detailed information about the commit (author, changed files/passwords etc.)

"Commit log" (GitLogActivity) on a phone:

Action bar menu with "Show commit log" (on Git repo):

"Commit log" (GitLogActivity) on a tablet:

_Suggestion_ (not included in the original state of this PR) for a row layout, in case the commit author's name should be included, too (possibly with a smaller message font):


app/src/main/java/com/zeapo/pwdstore/git/log/GitLogAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+
+        fun bind(commit: GitCommit) {
+            binding.apply {
+                gitLogRowMessage.text = commit.shortMessage
+                gitLogRowHash.text = shortHash(commit.hash)
+                gitLogRowTime.text = stringFrom(commit.time)
+            }
+        }
+    }
+}
```
fmeum: Nit: We try to have trailing newlines in every file.
```suggestion
}

```

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogModel.kt
```diff
@@ -0,0 +1,61 @@
...
+            GitCommit(hash(it), it.shortMessage, it.authorIdent.name, date(it))
+        }.toMutableList()
+    }
+    val size = cache.size
+
+    fun get(index: Int): GitCommit? {
+        if (index >= size) e { "Cannot get git commit with index $index. There are only $size." }
+        return cache.getOrNull(index)
+    }
+}
```
fmeum:
```suggestion
}

```

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogActivity.kt
```diff
@@ -0,0 +1,47 @@
...
+        viewAdapter = GitLogAdapter()
+        val viewItemDecoration = DividerItemDecoration(this, LinearLayoutManager.VERTICAL)
+        view = findViewById<RecyclerView>(R.id.git_log_recycler_view).apply {
+            setHasFixedSize(true)
+            addItemDecoration(viewItemDecoration)
+            layoutManager = viewLayoutManager
+            adapter = viewAdapter
+        }
+    }
+}
```
fmeum:
```suggestion
}

```

CHANGELOG.md
```diff
@@ -25,6 +26,10 @@ All notable changes to this project will be documented in this file.
 -   Saving a password after creating it fails to finish commit operation
 -   HTTPS authentication did not prompt users for password
 
+### Fixed
+
+-   Password creation UI will scroll if it does not fit on the screen
```
fmeum: @msfjarvis Not sure why this is shown in the diff, did it get lost when we released 1.11.2?

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogModel.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.github.ajalt.timberkt.e
+import com.zeapo.pwdstore.utils.PasswordRepository
+import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.lib.ObjectId
+import org.eclipse.jgit.revwalk.RevCommit
+import java.util.Date
+
+private fun hash(commit: RevCommit) = ObjectId.toString(commit.id)
+
+private fun date(commit: RevCommit): Date {
```
fmeum: I would prefer to have these be Kotlin extension properties of `RevCommit`. This would make it easier to reuse them in other contexts where the commit might be null and would convey more clearly that the returned values are obtained from simple manipulations of existing `RevCommit` properties.

app/src/main/res/menu/main_menu_git.xml
```diff
@@ -29,6 +29,10 @@
         android:id="@+id/refresh"
         android:title="@string/refresh_list" />
 
+    <item
```
fmeum: Please also add this item to `main_menu_no_auth.xml`, which is our menu for configurations without write access to the GIt repository.

fmeum (commented):
Thank you very much for the contribution, this looks pretty good and useful to me!

I only left a few comments that are mostly nits.

CHANGELOG.md
msfjarvis: Nope, it's there in the unreleased section. Seems like a rebase error.

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogActivity.kt
```diff
@@ -0,0 +1,47 @@
...
+        setContentView(binding.root)
+        supportActionBar?.setDisplayHomeAsUpEnabled(true)
+        createRecyclerView()
+    }
+
+    private fun createRecyclerView() {
+        viewLayoutManager = LinearLayoutManager(this)
+        viewAdapter = GitLogAdapter()
+        val viewItemDecoration = DividerItemDecoration(this, LinearLayoutManager.VERTICAL)
+        view = findViewById<RecyclerView>(R.id.git_log_recycler_view).apply {
```
msfjarvis:
```suggestion
        view = binding.gitLogRecyclerView.apply {
```

If you're using viewbinding then its preferable to use it throughout.

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+        }
+        viewHolder.bind(commit)
+    }
+
+    override fun getItemCount() = model.size
+
+    class ViewHolder(private val binding: GitLogRowLayoutBinding) : RecyclerView.ViewHolder(binding.root) {
+
+        fun bind(commit: GitCommit) {
+            binding.apply {
```
msfjarvis:
```suggestion
        fun bind(commit: GitCommit) = with(binding) {
```

This reads a bit cleaner IMO

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogActivity.kt
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * Displays the repository's git commits in git-log fashion.
+ *
+ * It provides basic information about each commit by way of a non-interactive RecyclerView.
+ */
+class GitLogActivity : BaseGitActivity() {
+
+    private val binding by viewBinding(ActivityGitLogBinding::inflate)
+    private lateinit var view: RecyclerView
+    private lateinit var viewLayoutManager: RecyclerView.LayoutManager
```
msfjarvis: rather than set the layout manager here, you can do it in the XML layout itself.

app/src/main/res/layout/activity_git_log.xml
```diff
@@ -0,0 +1,22 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="match_parent">
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/git_log_recycler_view"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:scrollbars="vertical"
+        tools:context="com.zeapo.pwdstore.git.log.GitLogActivity"
+        tools:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
```
msfjarvis:
```suggestion
        app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
```

This will set the layout manager directly, making it work for both the preview as well as the actual app.

app/src/main/res/menu/main_menu_git.xml
msfjarvis: The overflow menu is getting super crowded IMO so rather than add it here, it'd make much more sense to me if we get rid of the top hash from the Git utils screen and add a button to open the log from there.

msfjarvis (changes requested):
This is looking _extremely_ good! I just have a couple nits here and there.

app/src/main/res/menu/main_menu_git.xml
fmeum: Agreed, that's even better. Viewing the Git log is likely not something one does often enough to warrant inclusion in this menu. The Git utils screen seems appropriate, we might even want to rename it into something like "Git log & utils".

msfjarvis:
I had some time to spare while checking this out so I've addressed the simpler stuff, leaving you more time to focus on the more involved aspects. Let me know if you have questions about any of the changes I made.

app/src/main/res/menu/main_menu_git.xml
Nosweh: Thanks to both of you for your kind feedback! I'll gladly make the remaining changes.

I agree that the GitConfigActivity will probably be a more suitable place to access the log from. However, wouldn't you say that dropping the HEAD's commit hash there, which is accompanied either by the name of the current branch or the notice that HEAD is currently detached, would lose the user this information about the state of HEAD? Should we then keep it there or maybe provide this info another way?

If we keep it, how about changing the 'Commit hash 1234abcd (refs/heads/master [or DETACHED])' to be more direct Git wording - something like what you get on git-status: Either
- 'HEAD detached at 1234abcd' or
- 'On branch [refs/heads/]master'?

(Just an idea, though.)

msfjarvis: I think you're headed in the right direction with your idea, so please go ahead and implement that. Also while at it, either drop the 'Hackish tools' label or just rename it to something more sensible. I think these have been improved and rewritten enough to not be 'hacks' anymore.

Nosweh:
I have now made the commit log accessible via the GitConfigActivity, changed the HEAD info there to use git-status wording and changed the "Hackish tools" label to "Utilities" as well as the "Git utils" label to "Local Git config & utilities". I am sure the last label is not ideal, but then again, I could not think of a shorter label that would give enough information about the GitConfigActivity for someone who is searching for specific functionality and is unfamiliar with the app.

I tested the HEAD info once with HEAD pointing to master and once with HEAD pointing to a specific commit directly.

I also added the two extensions to RevCommit you asked for, @FabianHenneke .

app/src/main/java/com/zeapo/pwdstore/git/GitConfigActivity.kt
```diff
@@ -66,4 +54,51 @@ class GitConfigActivity : BaseGitActivity() {
             }
         }
     }
+
+    /**
+     * Sets up the UI components of the tools section.
+     */
+    private fun setupTools() {
+        val repoDir = PasswordRepository.getRepositoryDirectory()
+        val repo = PasswordRepository.getRepository(repoDir)
```
fmeum: Blame is on us, but this can actually be "simplified" to:
```suggestion
        val repo = PasswordRepository.getRepository(null)
```

app/src/main/java/com/zeapo/pwdstore/git/log/GitLogModel.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.github.ajalt.timberkt.e
+import com.zeapo.pwdstore.utils.PasswordRepository
+import com.zeapo.pwdstore.utils.hash
+import com.zeapo.pwdstore.utils.time
+import org.eclipse.jgit.api.Git
+import org.eclipse.jgit.revwalk.RevCommit
+
+private fun commits(): Iterable<RevCommit> {
+    val repoDir = PasswordRepository.getRepositoryDirectory()
+    val repo = PasswordRepository.getRepository(repoDir)
```
fmeum: See above
```suggestion
    val repo = PasswordRepository.getRepository(null)
```

fmeum (dismissed):
Looks great, works great, thanks a lot!

I only left two suggested changes that will help us when we finally get to refactor `PasswordRepository`.

app/src/main/res/values-es/strings.xml
```diff
@@ -178,7 +178,6 @@
     <string name="git_push_generic_error">El envío fue rechazado por el servidor, la razón:</string>
     <string name="jgit_error_push_dialog_text">Ocurrió un error durante el envío:</string>
     <string name="git_operation_remember_passphrase">Recordar contraseñagit  (inseguro)</string>
-    <string name="hackish_tools">Hackish tools</string>
+    <string name="git_tools">Hackish tools</string>
```
msfjarvis: Drop all the translations for `git_tools` since they're outdated now.

msfjarvis (dismissed):

fmeum (approved):

Nosweh:
Alright, should be all done now.

msfjarvis (approved):

msfjarvis:
Congratulations! :confetti_ball:

Nosweh:
Thanks, I'm glad I could contribute. And thanks for your support!

PR #1060, opened 2020-08-27
Delete HTTPS instead of SSH key password on error

fmeum:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## :scroll: Description

Since #995 (more specifically [this line](https://github.com/android-password-store/Android-Password-Store/pull/995/files#diff-dd33817c7095cf708bd0f776626e3738L160)), we are removing the SSH key passphrase instead of the HTTPS password. This is just plain wrong and appears to have slipped in during a merge.

Verified that the stored HTTPS password is removed when it causes an error due to failed authentication.

## :pencil: Checklist


msfjarvis (approved):
:disappointed:

Issue #1067, opened 2020-08-29
[BUG] Switching apps while running git operation causes crash


xdevs23:
**Describe the bug**
Switching away from APS while it is running a git operation (let's say, synchronizing repository) may cause a crash.
This can be reproduced if the app being switched to is **not** a launcher.

**To Reproduce**
1. Open a different app
2. Open APS
3. Tap three dots
4. Tap on "Synchronize repository"
5. Immediately switch to the other app without closing APS
7. Observe crash

**Expected behavior**
Does not crash after switch

**Screenshots**
Not posting publicly for privacy reasons.

**Device information:**
 - Device: OnePlus 5
 - OS: halogenOS 10 (AOSP 10 r41)
 - App version 1.12-free from dl.msfjarvis.dev, downloaded on 2020-08-29

**Additional context**
```
android.view.WindowManager$BadTokenException: Unable to add window -- token android.os.BinderProxy@a6abdd6 is not valid; is your activity running?
	at android.view.ViewRootImpl.setView(ViewRootImpl.java:931)
	at android.view.WindowManagerGlobal.addView(WindowManagerGlobal.java:387)
	at android.view.WindowManagerImpl.addView(WindowManagerImpl.java:95)
	at android.app.Dialog.show(Dialog.java:342)
	at androidx.appcompat.app.AlertDialog$Builder.show(AlertDialog.java:2)
	at com.zeapo.pwdstore.git.BaseGitActivity.launchGitOperation(BaseGitActivity.kt:28)
	at com.zeapo.pwdstore.git.BaseGitActivity$launchGitOperation$1.invokeSuspend(Unknown Source:12)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:13)
	at android.os.Handler.handleCallback(Handler.java:883)
	at android.os.Handler.dispatchMessage(Handler.java:100)
	at android.os.Looper.loop(Looper.java:214)
	at android.app.ActivityThread.main(ActivityThread.java:7356)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:925)
```

Biometric Auth (Fingerprint) is enabled.

[...]

PR #1070, opened 2020-08-31
Add Keystore backend for SSH public key authentication

fmeum:

- [ ] Bugfix
- [x] New feature
- [ ] Enhancement
- [x] Refactoring
- [ ] Dependency updates

## :scroll: Description

Implement an SSH auth method backed by an Android Keystore key and make SshKeyGenFragment generate such a key.

In detail, this means the following:

* SSH keys generated within the app are no longer kept in a simple file optionally encrypted with a passphrase, but rather in the Android Keystore optionally protected with the user's screen lock credential. On modern devices with a TEE this means that the private key will not be leaked even if the device is rooted or compromised remotely.
* SSH key files with passphrases remain supported for users who prefer them, they just can't be generated from within the app anymore.
* The offered key types are changed from RSA-2048 and RSA-4096 (default) to RSA-3072, P-256 and Ed25519 for the following reasons:
    * RSA-3072 is supported by all servers and provides the target security level of 128 bits.
    * P-256 is supported by most servers (support added to OpenSSH in 2011, will be in all major LTS distros by September) and provides fast authentication. Should maybe be the default since its supported natively by the Android Keystore so that we would never have the key material in memory in the context of our app. On modern devices, this key type uses the Strongbox.
    * Ed25519 since its what everybody should use if they can (support introduced in 2014, but should also be in all major LTS distros by September). However, it's not supported natively by the Keystore and we need to load the private key material from an encrypted file during authentication. Is the default for now, but we can talk about this.

The new `SshKey` objects deals with everything SSH pubkey related, e.g. also with key import. I also modernized the SSH key gen and key view activities.

Also, Small changes were made to BiometricAuthenticator to fix one of two memory leaks.

Supersedes #807.

Tried to authenticate with all key types. Please test this extensively.

## :pencil: Checklist

## :crystal_ball: Next steps

CHANGELOG entry


msfjarvis:
We're definitely going to have to document this in the wiki when it ships. I'll review now.

app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt
```diff
@@ -0,0 +1,321 @@
...
+            build()
+        }
+    }
+
+    @Suppress("BlockingMethodInNonBlockingContext")
+    suspend fun generateKeystoreWrappedEd25519Key(requireAuthentication: Boolean) = withContext(Dispatchers.IO) {
+        delete()
+
+        val encryptedPrivateKeyFile = getOrCreateWrappedPrivateKeyFile(requireAuthentication)
+        // Generate the ed255519 key pair and encrypt the private key.
```
msfjarvis: nit: ed25519

app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt
```diff
@@ -0,0 +1,321 @@
...
+        Type.KeystoreNative -> KeystoreNativeKeyProvider
+        Type.KeystoreWrappedEd25519 -> KeystoreWrappedEd25519KeyProvider
+        null -> null
+    }
+
+    private object KeystoreNativeKeyProvider : KeyProvider {
+
+        override fun getPublic(): PublicKey = try {
+            androidKeystore.sshPublicKey!!
+        } catch (error: Exception) {
```
msfjarvis:
```suggestion
        } catch (error: Throwable) {
```

If we're doing a catch-all then this makes more sense.

app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt
```diff
@@ -0,0 +1,321 @@
...
+            e(error)
+            throw IOException("Failed to get public key '$KEYSTORE_ALIAS' from Android Keystore", error)
+        }
+
+        override fun getPrivate(): PrivateKey = try {
+            androidKeystore.sshPrivateKey!!
+        } catch (error: Exception) {
```
msfjarvis: Same as above

app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+        override fun getType(): KeyType = KeyType.fromKey(public)
+    }
+
+    private object KeystoreWrappedEd25519KeyProvider : KeyProvider {
+
+        override fun getPublic(): PublicKey = try {
+            parseSshPublicKey(sshPublicKey!!)!!
+        } catch (error: Exception) {
```
msfjarvis: Same as above

app/src/main/java/com/zeapo/pwdstore/git/sshj/SshKey.kt
```diff
@@ -0,0 +1,321 @@
...
+        override fun getPrivate(): PrivateKey = try {
+            // The current MasterKey API does not allow getting a reference to an existing one
+            // without specifying the KeySpec for a new one. However, the value for passed here
+            // for `requireAuthentication` is not used as the key already exists at this point.
+            val encryptedPrivateKeyFile = runBlocking {
+                getOrCreateWrappedPrivateKeyFile(false)
+            }
+            val rawPrivateKey = encryptedPrivateKeyFile.openFileInput().use { it.readBytes() }
+            EdDSAPrivateKey(EdDSAPrivateKeySpec(rawPrivateKey, EdDSANamedCurveTable.ED_25519_CURVE_SPEC))
+        } catch (error: Exception) {
```
msfjarvis: Same as above

app/src/main/java/com/zeapo/pwdstore/sshkeygen/ShowSshKeyFragment.kt
```diff
@@ -4,61 +4,36 @@
...
-    private lateinit var builder: MaterialAlertDialogBuilder
-    private lateinit var publicKey: TextView
-
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        builder = MaterialAlertDialogBuilder(requireActivity())
-    }
-
-    @SuppressLint("InflateParams")
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
```
msfjarvis, line 17: Can we get rid of this entire fragment? Feels like doing it in the activity itself works just as well.

msfjarvis (commented):
Everything looks pretty solid, really great work here! I'll take it for a spin.

app/src/main/java/com/zeapo/pwdstore/sshkeygen/ShowSshKeyFragment.kt
fmeum: It's shown in two different places though (keygen and preferences) and putting the code next to either would feel out of place. Should we make it a top-level function instead and just show the dialog in the two places?

msfjarvis: Ah, I completely forgot about the preference case. Let's leave this as-is.

msfjarvis (approved):
Everything works in my testing, if the changelog entries look good then you can go ahead and merge it.

fmeum:
Thanks, they do a pretty good job at communicating the changes.

msfjarvis:
Is this change compatible for users upgrading from v1.11.3 w/ SSH key authentication? I privately contacted the reporter for #1067 to see if #1066 solved his problem but after upgrading to the latest snapshot git operations prompted that their SSH key is missing. Generating a new key fixed the issue, but we'll either need to add a migration or notify users that they'll need new SSH keys and why.

<details>
<summary>Screenshot</summary>

image


</details>

fmeum:
> Is this change compatible for users upgrading from v1.11.3 w/ SSH key authentication? I privately contacted the reporter for #1067 to see if #1066 solved his problem but after upgrading to the latest snapshot git operations prompted that their SSH key is missing. Generating a new key fixed the issue, but we'll either need to add a migration or notify users that they'll need new SSH keys and why.
> 
> Screenshot

Sorry about that, I forgot to add the (trivial) migration. Will send a PR in a minute.

PR #1073, opened 2020-09-03
Prevent double tap events during password selection

fmeum:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## :scroll: Description

The upstream bug that led to failed selections of password items right under each other when tapped in quick succession has now been fixed.

I would have said that this is too minor to warrant inclusion in the changelog, but am open to add it anyway.

Selection works much better now.

## :pencil: Checklist


msfjarvis (approved):
I saw that the issue was fixed a few days ago, then saw yesterday the artifact was updated, yet failed to put 2 and 2 together and update it here :roll_eyes: 

I don't think a changelog entry is necessary

PR #1138, opened 2020-10-06
Show new FillResponse right after publisher reset

fmeum:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring
- [ ] Dependency updates

## :scroll: Description

If Autofill shows a warning about an app whose publisher changed and the
user decides to trust the app and clear previous matches, they should
immediately be given the option to select a new match.

Previously, as AutofillPublisherChangedActivity did not return a
result, the old FillResponse with just a warning would be reused. We
now pass a useful response with no matches on to the activity, which
returns it after the user has chosen to reset the publisher info.

Resolves https://github.com/android-password-store/Android-Password-Store/pull/1136#issuecomment-703851026. Arguably this is not a workaround, but how Autofill intends this to be done.

Verified that the Autofill UI offers to search for a password entry right after the reset.

## :pencil: Checklist


msfjarvis (approved):
Confirmed working, thanks for the quickfix!

PR #3051, opened 2024-05-13
Update BasePGPActivity.kt with Clip extra sensitive description

eskalVAR:
My passwords were shown as plaintext in the "snackbar" this change fixes the issue.

This small change makes use of the "new" recommendation for sensitive data: https://developer.android.com/develop/ui/views/touch-and-input/copy-paste

Backward SDK level compatibility should still work.

Change has been tested on device emulator with Android 14.0 API 34

### DIFF
--- a/app/src/main/java/app/passwordstore/ui/crypto/BasePGPActivity.kt
+++ b/app/src/main/java/app/passwordstore/ui/crypto/BasePGPActivity.kt
@@ -6,6 +6,7 @@
 package app.passwordstore.ui.crypto
 
 import android.content.ClipData
+import android.content.ClipDescription
 import android.content.Intent
 import android.content.SharedPreferences
 import android.os.Build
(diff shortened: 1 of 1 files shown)


msfjarvis (changes requested):
`ClipDescription.EXTRA_IS_SENSITIVE` and `"android.content.extra.IS_SENSITIVE"` are the exact same thing, I don't see how this change actually affects anything. The existing code does the job correctly on my Pixel 8.

msfjarvis:
I just copied a password and took this screenshot, the password is correctly redacted (the background is black due to `FLAG_SECURE` set on `DecryptActivity`)

eskalVAR:
Huh interesting. My password is clear text with latest GrapheneOS build (2024050700) on Pixel 7. No issues with Aegis which after a quick look at the source uses the ClipDescription. Let me try the newly built APK on the actual device.

eskalVAR:
Newly built APK shows a redacted password. Version with clear text is latest available from F-Droid.

eskalVAR:
So even without change seems to be fine, my bad for not checking a fresh apk. Possibly feature was not added in release on f-droid?

msfjarvis:
> So even without change seems to be fine, my bad for not checking a fresh apk. Possibly feature was not added in release on f-droid?

There hasn't been a new release in a while so the code in `develop` is far ahead of the latest F-Droid and Google Play builds. [Snapshot builds](https://docs.passwordstore.app/docs/users/release-channels/) are provided for users who wish to use the newer code to aid with testing and stabilization.

eskalVAR:
Thanks for this. Just checked the release history and the change is not present even in in the latest V1.x tag.

Will try out a snap release.

-------- Original Message --------

